PR #2, opened 2019-07-28
Completed Competitive Coding - 4 - Problem 1

bharatreddypothireddy:
Implemented by finding the middle node and reversing the list after it and traversing both the lists.


ashmeetdeol19:
Well Done with both the solution Bharat ?

PR #7, opened 2019-07-28
Please review : Competitive coding 4 - First try

MahiteeS:
…t - first try

### DIFF
--- a/BalancedBinaryTree.java
+++ b/BalancedBinaryTree.java
@@ -1,0 +1,41 @@
+package competitiveCoding4;
+
+
+//Time COmplexity: O(n^2) because nodes of trees are traversed twice.
+// Space Complexity: O(1) because we use constant space
+public class BalancedBinaryTree {
+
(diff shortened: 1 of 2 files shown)


BalancedBinaryTree.java
```diff
@@ -0,0 +1,41 @@
+package competitiveCoding4;
+
+
+//Time COmplexity: O(n^2) because nodes of trees are traversed twice.
```
ashmeetdeol19: If nodes are traversed twice then complexity should be 2n.
Think about it again.

PalindromeLinkedLIst.java
```diff
@@ -0,0 +1,34 @@
...
+        
+        while (head != null) {
+            if (head.val != (int)stack.pop()) {
+                return false;
+            }
+            head = head.next;
+        }
+        return true;
+    }
+}
```
ashmeetdeol19: This can be optimized further ... Think

PR #50, opened 2020-04-14
Program1

AMAbhi:


Program1.py
```diff
@@ -0,0 +1,33 @@
+# Definition for singly-linked list.
+# class ListNode:
+#     def __init__(self, x):
+#         self.val = x
+#         self.next = None
+
+#Time Complexity - O(n)
+#Space Complexity - O(1)
```
devashishnyati, line 8: You are using extra space (stack). Try to optimize it.

devashishnyati:
Good job Abhishek!
Try to solve the problem without extra space. 
Submit the other problem as well.

PR #52, opened 2020-04-14
Please review

megrao:


balanced_binaryTree.py
```diff
@@ -0,0 +1,36 @@
+'''
+Executed in leetcode: Yes
+Time complexity: O(n log n) in worst case (skewed tree)?
```
devashishnyati, line 3: why nlogn? You are touching each node once. It should be O(n)

devashishnyati:
Good job Megha!
Try the iterative solution also problem 2.

PR #66, opened 2020-05-19
Update sample.java. Please review Competetive-Coding-4

Abinashi7:


iivaishnavii:
Good Job, Abhinash.
For Linked List problem, try also an constant space solution.
Hint: Slow and fast pointer

PR #74, opened 2020-05-26
completed competitive coding 4 -- Please review

VenkateswaraRajuDatla:


iivaishnavii:
Good job! You got your solution, TC and SC right

PR #83, opened 2020-06-22
Competetive-Coding-4 done

Akash-Akki:


AbhiniveshP:
Nice work!!!
You can simultaneously return boolean along with height in parallel which would be even more faster compared to the solution provided.

AbhiniveshP:
Nice work!!!
You can simultaneously return boolean along with height in parallel which would be even more faster compared to the solution provided.

PR #95, opened 2020-06-23
Completed Competitive-Coding-4 : Please review.

shantanu609:


AbhiniveshP:
Nice work Shantanu!!!
For balanced binary tree question, you can implement in O(N) time if you return height and boolean values from the same function instead of an extra function for height. Try implementing that.

AbhiniveshP:
Nice work Shantanu!!!
For balanced binary tree question, you can implement in O(N) time if you return height and boolean values from the same function instead of an extra function for height. Try implementing that.

shantanu609:
Updated the Problem2 with O(n) time solution.

PR #155, opened 2020-10-11
Palindrome linkedlist

DhruvJawalkar:


PalindromeLinkedList.java
```diff
@@ -0,0 +1,59 @@
+// Time Complexity : O(n)
+// Space Complexity : O(1)
+// Did this code successfully run on Leetcode : yes
+// Any problem you faced while coding this : where to chop the linkedlist in half, setting node.next to null
```
kannan-mrugapphan, line 4: The point to split is the middle because you have to check for equality of 1st half and reverse of second half. The other points are implementation details.

kannan-mrugapphan (commented):
Good Job. Please submit the other problem as well.

kannan-mrugapphan:
Good Job. Please submit the other problem as well.

PR #175, opened 2020-10-18
Create Problem1.py

siddhesh25:


Problem1.py
```diff
@@ -0,0 +1,40 @@
+# Definition for singly-linked list.
```
Lohith-Aswathappa, line 1: Include time and space complexities

Lohith-Aswathappa (commented):
Good work, Siddhesh!
Try the other problem as well.

Lohith-Aswathappa:
Good work, Siddhesh!
Try the other problem as well.

PR #179, opened 2020-11-09
CC4 done - Please review

GayatriHungund81194:


balancedBinarytree.py
```diff
@@ -0,0 +1,14 @@
+# The main concept here is that we need to see if height of left and right subtree diffrs by at most 1 so calculate recursively height of all left and right subtrees and then aadd 1 to it, then consioder the heoght as max of left and right subtree height
+# Time Complexity: O(h)
```
samarthjj, line 2: Time: O(n^2). The tree is completely traversed at each node to calculate height.
Please also try the bottom up approach which is more optimal.

samarthjj:
Good work here Gayatri, Please review the inline comments.

samarthjj:
Good work here Gayatri, Please review the inline comments.

PR #202, opened 2020-12-15
Problems Completed-Please Review

DivyaPuraswani:


kannan-mrugapphan (commented):
Good Job. Please try to optimize both problems. The first problem runs at O(n^2) time, try to get it down to O(n). The second problem can be done in O(1) space.

kannan-mrugapphan:
Good Job. Please try to optimize both problems. The first problem runs at O(n^2) time, try to get it down to O(n). The second problem can be done in O(1) space.

PR #204, opened 2020-12-16
Competitive-Coding-4 Complete. Please review. Thank you

Vasundhara-Krishna:


Balanced-BT.cpp
```diff
@@ -0,0 +1,35 @@
+// Time Complexity : O(nlogn)
```
samarthjj, line 1: Time: O(n^2). Height is recalculated for every node, tree traversed repeatedly. Please try the bottom up approach as well.

samarthjj:
Good job here Vasundhara, Please review the inline comments.

samarthjj:
Good job here Vasundhara, Please review the inline comments.

PR #268, opened 2021-03-16
Create BalancedBinaryTree.java

aswathimanden:


BalancedBinaryTree.java
```diff
@@ -0,0 +1,26 @@
+// Time Complexity : O(n)
+// Space Complexity : O(n)
+// Did this code successfully run on Leetcode : Yes 
+// Any problem you faced while coding this : No
+
```
sibalnirbhay, line 5: SC would b O(h) (where h = the height of the tree).

sibalnirbhay (commented):
Read inline comments and submit the other problem as well.

sibalnirbhay:
Read inline comments and submit the other problem as well.

PR #315, opened 2021-06-22
Completed CC- 4, please review

harshaGitHubDev:


Rooppesh:
Good Job Harsha! The TC for isBalanced would be O(n) since you would be visiting the nodes only once.

Rooppesh:
Good Job Harsha! The TC for isBalanced would be O(n) since you would be visiting the nodes only once.

PR #522, opened 2022-05-23
Implemented the algos for BT if Balanced, Linked List if Palindrome c…

arshmohd:
Implemented the algos for 
Problem#1:
---------
Determine if a given singly linked list is palindrome or not?

Problem#2:
-----------
You are given a binary tree find if it is height-balanced or not. for this problem assume that height -balanced tree is defined as:
a binary tree in which the depth of the two subtrees of every node never differ by more than 1.


Sumesh-H:
Well done Irfan!

PR #561, opened 2022-06-22
Update sample.java

darshil1899:


jeelpatel18 (commented):
Add TC and SC and also upload the 2nd problem as well.

jeelpatel18:
Add TC and SC and also upload the 2nd problem as well.

PR #693, opened 2023-06-23
completed compitative coding 4

Sailaja-ASB:


problem2.py
```diff
@@ -0,0 +1,33 @@
+"""
+Time Complexity : O(N)
+Space Complexity :O(N)
```
ishika2741, line 3: To be more concise,i t will be O(H) where h is the height of the tree which when given a skewed tree can become O(N) because N and H there will be same

ishika2741 (commented):
Well done! Keep it up.

ishika2741:
Well done! Keep it up.

PR #739, opened 2023-10-23
added both solutions

merinjputhiyedath:


keshav-Gupta-IS (commented):
Good Work, Merin!! Keep it up. Coming up with edge cases comes as we practice more and more.

keshav-Gupta-IS:
Good Work, Merin!! Keep it up. Coming up with edge cases comes as we practice more and more.

PR #810, opened 2024-04-05
Added Problem1.java

ZeelSatasiya:


ishika2741 (commented):
Kindly dd tc and sc as well. Also submit second problem.

ishika2741:
Kindly add tc and sc as well. Also submit second problem.

PR #1011, opened 2025-06-13
completed

IshaaS:
solved palindrome of linked list by finding middle, reversing second half and comparing.
solved balanced binary tree using dfs and comparing heights of left sub trees with right sub trees.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,3 +1,6 @@
 # Competitive-Coding-4
 
 Please submit the interview problems posted in slack channel here. The problems and statements are intentionally not shown here so that students are not able to see them in advance 
+
+problem1 http://leetcode.com/problems/balanced-binary-tree/description/
+problem2 https://leetcode.com/problems/palindrome-linked-list/description/
--- a/problem1.java
+++ b/problem1.java
@@ -1,0 +1,47 @@
+//http://leetcode.com/problems/balanced-binary-tree/description/
+
+// Time Complexity :O(n)
+// Space Complexity :O(h)
+// Did this code successfully run on Leetcode :yes
+// Any problem you faced while coding this :no
+
+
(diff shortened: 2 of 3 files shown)


super30admin:
Let's evaluate each problem solution separately:

Problem 1: Balanced Binary Tree
Strengths:
1. Correct approach using DFS to check subtree heights and balance.
2. Time complexity is correctly identified as O(n) where n is number of nodes.
3. Space complexity is correctly noted as O(h) where h is tree height (recursion stack).
4. Uses a helper function effectively to track depth.

Areas for Improvement:
1. The flag variable could be made local to the function rather than a class variable.
2. The depth comparison logic could be simplified to `if(Math.abs(lstDepth - rstDepth) > 1) flag = false;`
3. Could add null check at the start of isBalanced() for robustness.
4. The depth parameter in helper() starts at 1 which is unconventional (typically starts at 0).

Problem 2: Palindrome Linked List
Strengths:
1. Correct approach using fast-slow pointer technique to find middle.
2. Proper reversal of second half of list.
3. Correct time complexity O(n) and space complexity O(1).
4. Handles edge cases well (null or single node).

Areas for Improvement:
1. Variable names could be more descriptive (e.g., 'half' could be 'slow').
2. Could add comments to clearly separate the three main steps.
3. The reversal could be extracted into a separate helper function for better modularity.
4. The comparison loop could check firstHalf != null for symmetry, though not strictly necessary.

General Observations:
1. Both solutions show good understanding of the problems.
2. Time and space complexity analysis is accurate.
3. Code is generally clean and readable.
4. Could benefit from more consistent spacing (some dense blocks of code).
5. Comments are helpful but could be more detailed in explaining the approach.

PR #1050, opened 2025-09-16
Add solutions for height-balanced tree and palindrome linked list

ravithakur2k:
Added HeightBalancedTree.py with a function to check if a binary tree is height-balanced, and PalindromeLL.py with a function to check if a singly linked list is a palindrome. Both implementations include comments on time and space complexity.

### DIFF
--- a/HeightBalancedTree.py
+++ b/HeightBalancedTree.py
@@ -1,0 +1,30 @@
+# Time Complexity : O(n)
+# Space Complexity : O(n) recursive stack trace
+# Did this code successfully run on Leetcode : Yes
+# Any problem you faced while coding this : No
+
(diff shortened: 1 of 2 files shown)


super30admin:
Strengths:
- Clear and concise implementation for both problems.
- Good use of comments to explain the logic.
- Correct handling of edge cases (e.g., empty tree or single-node list).

Areas for Improvement:
- In the linked list solution, consider adding a check for an empty list at the beginning for robustness.
- For the tree solution, while using a class variable (`self.result`) works, it might be cleaner to return a tuple (height, isBalanced) from the helper function to avoid using class state.
- The comment in the tree solution mentions "value exceeds more than 1" which could be clearer as "height difference exceeds 1".

PR #1063, opened 2025-12-09

pranavs0133:


super30admin:
- Problem1: Well done! Your solution is efficient and correctly implements the palindrome check for a linked list. The approach of finding the middle, reversing the second half, and comparing is optimal.
- Problem2: Your solution works but can be optimized. Currently, you're calculating the height of subtrees multiple times, leading to O(n^2) time complexity. Consider a bottom-up approach where you calculate height and check balance in the same recursion, which would reduce it to O(n) time. Also, the global flag could be avoided by returning both height and balance status from the helper function.

PR #1074, opened 2026-02-02
1 problem completed

BharathVuppala96:


super30admin:
Your solution is on the right track and uses the optimal approach of reversing the second half. However, there are critical issues in handling the middle node and the list splitting.

Key issues:
1. You need to split the list into two halves. In the reference solution, after finding the middle (using slow and fast pointers), the list is split by setting `slow.next = None` before reversing the second half. This ensures that the first half is disconnected from the second half. Without this, the first half still includes the entire list up to the middle, and the reversal of the second half changes the pointers, but the first half's last node still points to the original next (which is now reversed). This leads to incorrect list structures for comparison.

2. For the reversal, you should start from the node after the middle. In the reference solution, the reversal is done on `slow.next` (which is the beginning of the second half). This is because when the list has an odd number of nodes, the middle node should be skipped during comparison. For example, in [1,2,3,2,1], the middle node is 3. We only compare [1,2] with [1,2] (the reversed second half). So you should reverse the list starting from `slow.next` instead of `slow`.

3. In the comparison step, you should only compare until the reversed list (second half) is exhausted. Since the first half might have one more node (in the odd case), but you broke the list so the first half has the same number of nodes as the reversed half? Actually, when you split correctly, the first half will have ceil(n/2) nodes and the reversed half will have floor(n/2) nodes. So for odd-length, the first half has one extra node (the middle). But during comparison, we only compare the first floor(n/2) nodes of the first half with the entire reversed half. So in the reference solution, the comparison loop runs while both halves are not null, but the reversed half has exactly floor(n/2) nodes. So when the reversed half is exhausted, we stop. The extra node in the first half (if any) is ignored.

How to fix:
- After finding the middle with slow and fast, you should reverse the list starting from `slow.next` (if you

PR #1075, opened 2026-02-07
Submitting Q1

akshay4121:


super30admin:
It seems there might have been a misunderstanding. The problem you solved is about checking if a binary tree is balanced, but the assigned problem is about checking if a singly linked list is a palindrome. These are two different problems.

For the linked list palindrome problem, you need to:
1. Find the middle of the linked list (using the slow and fast pointer technique).
2. Reverse the second half of the linked list.
4. Optionally, reverse the second half back to its original form (if required, though not necessary for the problem).

Your current solution does not address this. Please review the problem statement again and implement the solution for the correct problem.

If you intended to submit a solution for the balanced binary tree problem, note that the problem context here is for a linked list. Always ensure you are solving the correct problem as per the given title and description.

super30admin:
It seems there has been a confusion in the problem you are solving. The problem given is to check if a singly linked list is a palindrome, but your solution is for checking if a binary tree is balanced. Please make sure to read the problem statement carefully.

For the linked list palindrome problem, you need to:
1. Find the middle of the linked list (using slow and fast pointers).
2. Reverse the second half of the linked list.
4. Optionally, reverse the second half back to its original form (if required, though not necessary for the problem).

Your current solution for the balanced tree problem is correct for that problem, but it is not relevant here. I encourage you to revisit the problem statement and implement the solution for the linked list palindrome.

For the linked list problem, you should define a ListNode class and implement the isPalindrome method accordingly. The reference solution provided in the problem is a good starting point.

PR #1083, opened 2026-03-15
COmpetitive coding 4 doer

pranjay01:


super30admin:
### Interview Problem: is Linked List palindrome (Problem1.py)
It seems there was a confusion in the problem you were solving. The problem given was to check if a singly linked list is a palindrome, but your solution is for a binary tree balance check. Please ensure you are solving the correct problem.

For the linked list palindrome problem, you need to work with a linked list node structure (ListNode) instead of TreeNode. The approach typically involves:
1. Finding the middle of the linked list (using slow and fast pointers).
2. Reversing the second half of the list.
3. Comparing the first half and the reversed second half.
4. (Optional) Restoring the list by reversing the second half back.

Your current solution does not address any of these steps. I recommend revisiting the problem statement and implementing the correct algorithm for linked lists.

If you intended to submit a solution for the binary tree problem, please double-check the problem code and title to avoid such mix-ups.

VERDICT: NEEDS_IMPROVEMENT

---

### Interview Problem: Height Balanced Binary Tree  (Problem2.py)
It appears there has been a mix-up in the problem you were solving. The problem you were given is about determining if a binary tree is height-balanced, but your solution is for "234. Palindrome Linked List". 

For the correct problem (Height Balanced Binary Tree), you need to work with binary trees, not linked lists. A binary tree is height-balanced if the left and right subtrees of every node differ in height by no more than 1. 

A common approach is to use a recursive function that returns the height of the tree and also checks balance at each node. However, your current solution does not address this at all.

To solve the correct problem:
- You should define a function that calculates the height of a tree recursively.
- Additionally, you need to check at each node if the difference in heights of left and right subtrees is no more than 1.
- You can combine these checks in a single recursive pass to avoid O(n^2) time by returning both height and balance status.

For example, in Python:
```
def isBalanced(self, root: Optional[TreeNode]) -> bool:
    def check(node):
        if not node:
            return 0, True
        left_height, left_balanced = check(node.left)
        right_height, right_balanced = check(node.right)
        balanced = left_balanced and right_balanced and abs(left_height - right_height) <= 1
        return max(left_height, right_height) + 1, balanced
    return check(root)[1]
```
This approach runs in O(n) time and O(h) space (where h is the height of the tree), which is more efficient than the O(n^2) reference solution.

Please ensure you are solving the correct problem in the future. Double-check the problem statement and examples before coding.

VERDICT: NEEDS_IMPROVEMENT

pranjay01:
Please recheck properly, the issue is just the order Problem1.py has isBalanced solution and Problem2.py has isPalindrome solution

PR #1085, opened 2026-04-02
Add palindrome check for linked list

dewann:
Implement a solution to check if a linked list is a palindrome.

### DIFF
--- a/linkedlistpalindrome
+++ b/linkedlistpalindrome
@@ -1,0 +1,62 @@
+/**
+ * Definition for singly-linked list.
+ * public class ListNode {
+ *     int val;
+ *     ListNode next;
+ *     ListNode() {}
+ *     ListNode(int val) { this.val = val; }
+ *     ListNode(int val, ListNode next) { this.val = val; this.next = next; }
+ * }
+ */
+//tc On //sc O1
+class Solution {
+    public boolean isPalindrome(ListNode head) {
+        ListNode slow = head;
+
+        ListNode mid = findMid(slow);
+     
+        ListNode newhead = rev(mid);
+        mid.next = null;
+        mid = head;
+        return compareVals(mid, newhead);
+        
+
+    }
+
+    private boolean compareVals(ListNode start, ListNode end) {
+
+        while (null != start && null != end && start != end) {
+           
+            if ( start.val != end.val)
+               return false;
+            start = start.next;
+            end = end.next;
+        }
+        return true;
+    }
+
+    private ListNode rev(ListNode head) {
+        ListNode prev = null;
+        ListNode curr = head;
+        while (curr != null) {
+            ListNode temp = curr.next;
+            curr.next = prev;
+            prev = curr;
+            curr = temp;
+
+        }
+     
+        return prev;
+    }
+
+    private ListNode findMid(ListNode head) {
+        ListNode slow = head;
+        ListNode fast = head;
+        while (fast != null && fast.next != null) {
+            slow = slow.next;
+            fast = fast.next.next;
+
+        }
+        return slow;
+    }
+}


super30admin:
### Interview Problem: is Linked List palindrome (linkedlistpalindrome)
```

EVALUATION:
The student's solution attempts to solve the palindrome linked list problem by finding the middle node, reversing the second half, and then comparing the first half with the reversed second half. This approach is correct and meets the O(n) time and O(1) space requirements. However, there are a few issues in the implementation that need to be addressed.

1. **Correctness**: The solution has a logical error in the `compareVals` method. The condition `start != end` in the while loop is incorrect because after reversing, the second half's head (newhead) is not the same node as the middle node from the original list. Also, when the list has an odd number of nodes, the first half will have one more node than the second half. The comparison should only iterate through the length of the second half (reversed part). The current comparison might run into issues when the lists are of different lengths.

2. **Time Complexity**: The solution correctly has O(n) time complexity: finding the middle takes O(n), reversing the second half takes O(n), and comparing takes O(n). So overall O(n).

3. **Space Complexity**: The solution uses only constant extra space, so O(1), which meets the requirement.

4. **Code Quality**: The code is mostly readable and well-structured. However, variable names could be more descriptive (e.g., `mid` is used for both the middle node and then as the head of the first half). Also, the method `findMid` returns the middle node, but in the case of an even-length list, it returns the second middle node (which is correct for splitting). But note that when the list has an odd number of nodes, the middle node is exactly the center, and we should not include it in the second half? Actually, the common approach is to split the list at the middle, so for odd length, the center node is not compared (because it's the same). However, in the student's code, the `rev` method is called on the node returned by `findMid`, which is the middle node for odd-length lists. Then, when comparing, the first half starts at `head` and the second half starts at the reversed head. For an odd-length list, the first half has one more node (the center) than the second half. The comparison should stop when the second half is exhausted. But the student's `compareVals` method uses `start != end` which is not appropriate.

5. **Efficiency**: The approach is efficient, but the comparison function has a flaw. Also, the student sets `mid.next = null` to break the list, which is good.

Specifically, the issue in `compareVals`:
- The while loop condition `while (null != start && null != end && start != end)` is problematic. The `start != end` condition is not necessary and actually harmful. For example, if the list is [1,2,3,2,1], after reversing the second half (from node 3), the reversed list is [1,2,3]. Then we compare the first half [1,2,3] with the reversed second half [1,2,3]. But note the first half includes the center node (3) and the reversed second half also includes 3. So when we compare, we traverse both lists until one becomes null. However, the condition `start != end` would break when the nodes are the same? Actually, it breaks when the node references are the same, which might not happen. Instead, we should only compare until the reversed list (which is the second half) is exhausted. Because the first half might be longer (in odd-length case). So we should only check while `end != null` (since the reversed second half is shorter or equal? Actually, for even length, both halves are equal; for odd, the first half has one more). So the correct way is to traverse the reversed list until it becomes null.

Also, the student's code does not handle the case when the list is empty or has one node? Actually, the code should work for one node: `findMid` returns the head, then reversing it returns the same head, and then comparing head and head would return true. But in the comparison function, the while loop condition `start != end` might cause it to skip? For one node: start = head, end = head. Then the condition `start != end` is false, so the loop doesn't run, and it returns true. So it works. But for two nodes: [1,2]. The middle node (returned by findMid) is the second node. Then we reverse the second half (which is [2]) -> still [2]. Then we compare first half [1] and second half [2]. The while loop runs: start=1, end=2 -> they are not equal? So it returns false. Correct. But the condition `start != end` is true (because they are different nodes), so the loop runs. So it works in this case. However, if the list is [1,1], then the two nodes are the same. But when we compare, start=node1 (val=1), end=node2 (val=1). The nodes are different, so the loop runs. Then we compare values: equal. Then we move to next: start=node2, end=null. Then the condition fails (end is null) so we break and return true. So it works. But wait, when we have an odd-length list: [1,2,1]. The middle node is the node with value 2. Then we reverse the second half (starting from 2): reversing [2,1] becomes [1,2]. Then we set mid.next=null? Actually, in the code, after calling `findMid`, we get the middle node (node2). Then we call `rev(mid)`, which reverses the list starting from mid. So the reversed list is [1,2]. Then we set `mid.next = null`? This breaks the original list after the middle node. So the first half is [1,2] and the second half (reversed) is [1,2]. Then we compare: start=node1 (val1), end=node1' (val1) -> they are different nodes? So the loop runs. Compare: equal. Then next: start=node2 (val2), end=node2' (val2) -> again different nodes. Then compare: equal. Then next: start=null, end=null. So the loop ends and returns true. So it works.

But why does it work? Because the condition `start != end` is always true until they become null? Actually, the nodes from the first half and the reversed second half are distinct nodes (except in the case of a list with one node). So the condition `start != end` is redundant because it is always true until one becomes null. So it doesn't harm. However, it is unnecessary.

But there is a real problem: when the list has an even number of nodes, the two halves are equal in length. When it has an odd number, the first half is longer by one. The comparison should only compare the common length (i.e., the length of the reversed list). So the while loop should run as long as both start and end are not null. The condition `start != end` is not needed and might be confusing.

Another issue: in the code, the student writes `mid = head` after reversing, and then passes `mid` and `newhead` to compare. This is correct because `mid` is set to the head of the first half.

However, there is a critical error: when the list is empty? The problem states the number of nodes is at least 1. So we don't need to handle empty.

But what if the list has only one node? The code works as described.

So the code is actually correct? But wait, let's test with a list of two nodes that are equal: [1,1]. We already did and it worked.

But there is a potential issue: the `findMid` function. For even-length list, it returns the second middle. For example, [1,2,3,4] would return the node with value 3. Then we reverse from 3: [4,3]. Then we compare the first half [1,2] and the second half [4,3]. They are not equal. So it returns false. Correct.

So the code seems correct. Why did I think there was an issue? Actually, the condition `start != end` is harmless. But it is unnecessary and might be confusing. Also, the code does not handle the case when the reversed list is longer? It cannot be because we reversed exactly the second half.

But note: when we break the list by setting `mid.next = null`, the first half ends at the middle node. The reversed list starts from the node after the middle? Actually, no: the `findMid` returns the middle node. Then we reverse `mid` (which is the middle node) and beyond. So for odd-length, the middle node is included in the second half? But then when we break by setting `mid.next = null`, we are breaking the list at the middle node. So the first half includes the middle node? Actually, no: the first half is from head to the middle node (inclusive)? Because we did not change the first half. Then when we compare, we traverse the first half from head to the middle node (which is still there) and the reversed second half (which starts from the middle node? Actually, no: the reversed list starts from the last node of the original list).

VERDICT: NEEDS_IMPROVEMENT

---

### Interview Problem: Height Balanced Binary Tree 
It seems there was a misunderstanding: the problem you solved is for checking if a linked list is a palindrome, but the assigned problem is about determining if a binary tree is height-balanced. These are two different data structures and problems.

For the linked list palindrome solution you provided:
- The approach is correct: find the middle, reverse the second half, and compare.
- However, there are a few potential issues:
  1. In the `compareVals` method, the condition `start != end` might not be necessary. Actually, in this context, `start` and `end` are from two different halves of the list, so they are never the same node (unless the list has only one node). So this condition might be redundant. You can simply check that both are not null.
  2. For lists with an odd number of nodes, the middle node is not compared, which is correct. But ensure that the first half is correctly split. Currently, you set `mid.next = null` which breaks the list at the middle. Then the first half (starting at `head`) goes up to the node before the middle. This is correct.
  3. The method names could be more descriptive (e.g., `findMid` is good, but `rev` should be `reverse` for clarity).

For the actual problem (height-balanced binary tree):
- You need to work with a binary tree, not a linked list.
- The problem requires checking if the tree is balanced, meaning for every node the difference between the height of the left and right subtrees is at most 1.
- A common solution is to use a recursive function that returns the height of the tree and also checks balance. Alternatively, you can use a helper function that returns the height and check balance during the traversal.
- The reference solution provided has O(n^2) time complexity because it calculates height recursively for each node. There is a more efficient O(n) solution that checks balance while computing height.

Please review the problem statement again and implement the solution for the correct problem.

VERDICT: NEEDS_IMPROVEMENT

PR #1086, opened 2026-05-03
Completed competative coding 4

pratikb0501:


super30admin:
### Interview Problem: is Linked List palindrome (Problem_1.py)
It seems there was a mix-up in the problem you were solving. The problem asked is to check if a singly linked list is a palindrome, but your solution is for checking if a binary tree is balanced. Please ensure you are solving the correct problem.

For the linked list palindrome problem, you need to:
1. Find the middle of the linked list (using slow and fast pointers).
2. Reverse the second half of the list.
4. Optionally, reverse the second half back to restore the list (if required by the problem constraints, though it's not always necessary).

Your current solution does not handle linked lists at all. I recommend starting over with the correct problem in mind. Here is a step-by-step approach for the linked list palindrome problem:
- Use two pointers (slow and fast) to find the middle.
- Reverse the second half.
- Compare the first half and the reversed second half node by node.
- If all values match, return true; otherwise, false.

VERDICT: NEEDS_IMPROVEMENT

---

### Interview Problem: Height Balanced Binary Tree  (Problem_2.py)
It appears there has been a mix-up in the code submission. The problem you were asked to solve is about determining if a binary tree is height-balanced, but the code you provided is for checking if a linked list is a palindrome. 

For the Height Balanced Binary Tree problem, you need to work with binary trees, not linked lists. A height-balanced binary tree is one where the left and right subtrees of every node differ in height by no more than 1. 

I recommend reviewing the problem statement again and implementing a solution that:
1. Calculates the height of the tree (or subtrees) recursively.
2. Checks at each node if the absolute difference between the left and right subtree heights is at most 1.
3. Ensures that both the left and right subtrees are also balanced.

You can refer to the reference solution provided, which uses a recursive approach with a helper function to compute the height. However, note that the reference solution has a time complexity of O(n^2) due to repeated height calculations. You might consider optimizing this by combining the height calculation and balance check in a single recursive function that returns both the height and whether the subtree is balanced, which would bring the time complexity down to O(n).

VERDICT: NEEDS_IMPROVEMENT

